PR #26, opened 2025-01-08
Fix to allow spadesCBM to run all managed forests in SK

camillegiuliano:
Currently, spadesCBM runs all of SK in 1998. 
I am still getting errors running the following year (1999). I suspect this is due to a pixel group having 0 pixels and being removed when certain tables are merged. Trying to find a fix for this. 

Current iteration has also been tested with our smaller raster in 1998-2000 with consistent results. I also ran 1985-2011 with no issue.


camillegiuliano:
Problem in 1999 in the 1998-2000 run is due to the loss of a pixel group. There are no pixels in pixel group 974, and it disappears in most tables, this becomes a problem when dealing with `cbm_vars` where most tables just add the number of rows corresponding to the number of new pixel groups. This causes all tables in `cbm_vars` to be one row longer than other tables, causing errors. 

I need to put some thought into how to deal with this, I'm thinking the best solution will probably be to add some sort of check between the tables of `cbm_vars` and something like `pixelCount` to remove lost pixelGroups from these tables. Might be a bit harder for tables that don't have a `pixelGroup` or `row_idx` column.

PR #40, opened 2025-02-25
first year pixelGroup loss error fix and updating cbm_vars$parameter$age

camillegiuliano:
This fixes the error when pixelGroups disappear in the first simulation year (i.e. 1987 error). While fixing this error, found that cbm_vars$parameters$age was never updated from one year to the next, ages were static. Also included a fix for this, ages should now be updated yearly to match with those in cbm_vars$state$age

### DIFF
--- a/CBM_core.R
+++ b/CBM_core.R
@@ -628,6 +628,13 @@ annual <- function(sim) {
       cbm_vars$parameters$row_idx <- c(na.omit(cbm_vars$parameters$row_idx), newGCpixelGroup$pixelGroup)
     }
 
+    # Adding ages
+    if (is.null(cbm_vars$parameters$age)) {
+      cbm_vars$parameters$age <- c(cbm_vars$state$age, rep(1, length(unique(newGCpixelGroup$pixelGroup))))
+    } else {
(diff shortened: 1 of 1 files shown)


cboisvenue:
@camillegiuliano - ages are updated internally in the python step function
```
##CELINE NOTES Ages are update in cbm_vars$state internally in the
  ##libcbmr::cbm_exn_step function
  # 2. increment ages
  sim$pixelGroupC$ages <- cbm_vars$state$age
```
Can you check that your fix does not update the ages twice?

camillegiuliano:
Yes, but the python functions only updates the ages of `cbm_vars$state` and not those of `cbm_vars$parameters`. Meaning the values in the `age` column in that table never change once they are assigned (i.e. all new pixelGroups will forever have an `age` of 1 until the end of the simulation).
All my fix does is add this line below to assign the updated ages of `cbm_vars$state` from the python function to `cbm_vars$parameters` to make sure the `age` column in `cbm_vars$parameters` gets updated to the current value rather than stay static across ages once it is assigned. 
```
cbm_vars$parameters$age <- cbm_vars$state$age
```

Short answer: This change won't affect the two objects in your code chunk, it just ensures that the `cbm_vars$parameters` table stays up to date with them.

PR #41, opened 2025-02-28
mySpuDmids now has an optional "sw_hw" column

suz-estella:
`mySpuDmids` now has an optional "sw_hw" column where disturbance matrix IDs can be specific to SW or HW stands.

@cboisvenue I realized that this wasn't as simple as just updating the matrix IDs in the `mySpuDmids` table coming from `CBM_dataPrep_SK`. This is an attempt to have `CBM_core` be ready to accept our new  `mySpuDmids` made with the cbm_exn CSV. I'm curious to hear if this logic is sound to you!

Because of the new "sw_hw" column that is in the [cbm_exn disturbance_matrix_association table](https://github.com/cat-cfs/libcbm_py/blob/main/libcbm/resources/cbm_exn/disturbance_matrix_association.csv) and not in the defaults DB, there will now be 2 `disturbance_matrix_id` possible for every pixel combo of `spatial_unit_id` and `disturbance_type_id` - one for SW and one for HW. Before assigning the matrix IDs, each pixel is determined to be SW or HW based on the info in `sim$gcid_is_sw_hw`.

For backwards compatibility with the defaults DB, this will still work if the matrix ID is the same for both SW or HW.

`spadesCBM` test update will have to follow this -- will link in a comment.

### DIFF
--- a/CBM_core.R
+++ b/CBM_core.R
@@ -79,6 +79,7 @@ defineModule(sim, list(
       columns = c(
         rasterID              = "ID links to pixel values in the disturbance rasters",
         wholeStand            = "Specifies if the whole stand is disturbed (1 = TRUE; 0 = FALSE)",
+        sw_hw                 = "Optional column specifying if disturbance applies to SW or HW",
         spatial_unit_id       = "Spatial unit ID",
         disturbance_type_id   = "Disturbance type ID",
         disturbance_matrix_id = "Disturbance matrix ID",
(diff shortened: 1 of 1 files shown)


suz-estella:
this PR in `spadesCBM` also lists the other PRs related to this update: https://github.com/PredictiveEcology/spadesCBM/pull/18

cboisvenue:
@suz-estella @camillegiuliano we had another way of determining the flag for sw or hw here:
https://github.com/PredictiveEcology/CBM_core/blob/7f8a37a40715ea26640ad906ee4cf555925f584f/CBM_core.R#L292
We need to get rid of that as that flag will come from the column in the cbm_exn csv files.
I would rather we not keep any links or reference to the `disturbance_matrix_id` from the default DB. We will not be using those and we will not eventually go back to using those. We don't need backward compatibility here because we will always want to be able to track species (as opposed to sw versus hw like CBM-CFS3).
So, maybe we can review this PR with this in mind? Happy to talk about it more.

CBM_core.R
```diff
@@ -481,23 +482,34 @@ annual <- function(sim) {
...
+
+  ## Check that there is one disturbance_matrix_id for each disturbance setting
+  distUnq <- c("spatial_unit_id", "sw_hw", "events")
+  if (nrow(unique(mySpuDmids[, c(distUnq, "disturbance_matrix_id"), with = FALSE])) >
+      nrow(unique(mySpuDmids[, distUnq, with = FALSE]))) stop(
+        "'mySpuDmids' must have only 1 'disturbance_matrix_id' each combination of ",
+        paste(sQuote(distUnq), collapse = ", "))
+
+  gcidDist <- merge(
+    copy(sim$gcid_is_sw_hw)[, .(gcids, is_sw, sw_hw = ifelse(is_sw, "sw", "hw"))],
```
suz-estella, line 499: As this PR is now, this takes the flag already set in `sim$gcid_is_sw_hw` (created where you linked) and uses it as an additional column to help match the cbm_exn `disturbance_matrix_id` to our pixels, since it seems that we need to know if the pixel is SW or HW to choose the ID: https://github.com/cat-cfs/libcbm_py/blob/main/libcbm/resources/cbm_exn/disturbance_matrix_association.csv

However, maybe this isn't what we want to do exactly! If I'm way off, hopefully this can be a good starting point to an in-person review of it all next week. Maybe there's more that needs to be changed to finish this update?

regarding backwards compatibility: this does not rely on any old `disturbance_matrix_id` being in `mySpuDmids`, instead, it's just that `mySpuDmids` _could_ still be input without a "hw_sw" column (as if it is from the defaults DB) and this would still work. It would just assign the same `disturbance_matrix_id` to SW and HW (like it was doing previously). If we never need that functionality, that's ok! Really it's just the inclusion of this line: https://github.com/suz-estella/CBM_core/blob/07e9c7483b161ca044ca812ebb9c7708642a5d46/CBM_core.R#L488

cboisvenue: Thank you for the clarification Susan. Ok. I will accept and we can sort this out on Monday in person.

PR #42, opened 2025-03-10
Disturbances simplified: disturbances table now has 1 column per year

suz-estella:
Inspired by bringing the RIA modules back in. @cboisvenue am I correct in thinking that the data table input in there specifically to work with the RIA modules?

It seems as though you can pass a table with the columns "year" and "pixels" where the pixel column contains lists for each year containing the disturbance pixel values. I think it allows there to be multiple lists for different types of disturbances (e.g. one list of pixels for wildfire, one list for clearcut). I'm not entirely sure how to use it just by looking at it.

This is much simpler in that it expects there to be a column called "pixelIndex" and then a column for each year (.e.g. "2025") where the values would be the values in a disturbance raster. The only difference is that the different types of disturbances will need to be merged into a single column ahead of time, which is indeed a bit more work, but I think it makes understanding how to provide the input a bit easier.

What do you think? Otherwise, I could add the "disturbance-merging" process back in with a more detailed metadata description.

At least, I've made the code a bit more efficient here, where more of the disturbance-value wrangling process is shared for all the inputs.

### DIFF
--- a/CBM_core.R
+++ b/CBM_core.R
@@ -98,7 +98,7 @@ defineModule(sim, list(
         "(a character path to a raster file or a terra SpatRaster)",
         "can be accessed by subsetting the object with the 4 digit year name",
         "(e.g. sim$disturbanceRasters[[\"1990\"]]);",
-        "2. A data.table with the first column containing pixel values and another named 'year'."
+        "2. A data.table with a 'pixelIndex' column and additional columns named by disturbance year."
       )),
     expectsInput(
       objectName = "historicDMtype", objectClass = "numeric", sourceURL = NA,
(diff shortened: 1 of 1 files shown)


cboisvenue:
@suz-estella  this "we expect all the disturbances to be merged into a single raster layer" is not the case. Disturbances can be provided as a data.table with one column per disturbance. In RIA simulations (FRI scenario for example), disturbances are given in a data.table. 
I am not against having `CBM_core` be aspatial. But we have to remember that disturbances could be provided on a time-step basis, so not all at once prior to simulation. Think of running `spadesCBM` at the same time as `fireSense` or `scfm` and one of those two outputting a raster each year that needs to be taken into `CBM_core` on the fly to represent the disturbance of that year. So I don't see how that can't be processed in `CBM_core` since all the other modules' jobs are already done.
Let me know what you think. Happy to talk in person too!

suz-estella:
Edited my comment for clarity!
> we expect all the disturbances (_** if provided as a raster_) to be merged into a single raster layer, so it would be the most similar to also expect all the disturbances to be merged into a single column (**_if they are provided in a table_)

Definitely keeping in mind that we want to keep `CBM_core` open to providing disturbances on a yearly basis.

Of course my first thought is: would it be possible to have the modules take on another step to summarize those values into a table before they are sent off as input to `CBM_core`? I suppose the problem is that they don't have `masterRaster` as an input available to do this. But I imagine they also don't save the rasters with the name `disturbanceRasters`...? -- Makes me wonder: is there usually a "bridging" module in these scenarios that maps the inputs & outputs between the two (say between `scfm` and `CBM_core`)? I'll take a closer look at the RIA stuff and attempt to answer my own question!

suz-estella:
Closing this PR in favor of this one: https://github.com/PredictiveEcology/CBM_core/pull/44

PR #44, opened 2025-03-24
disturbanceEvents replaces disturbanceRasters; disturbanceMeta replaces mySpuDmids

suz-estella:
**NOTE: this is set as a draft because it will have to be merged at the same time as another PR to CBM_dataPrep_SK. I will link that here and mark all PRs for review at the same time.**

The purpose of this update is to standardize how disturbances are input to the `CBM_core` module to make it straightforward to link other modules to it. This moves `CBM_core` close to being entirely based on table (non-spatial) inputs (more to come there).

@cboisvenue interested in your thoughts on this!
@DominiqueCaron do you think this will make things easier for your LandR connections? Any recommendations?

Two major changes to the module inputs:
1. `disturbanceEvents` replaces `disturbanceRasters`

`disturbanceEvents` must always be a table with the columns "pixelIndex", "year", and "eventID". Any spatial processing (e.g. alignment with the master raster) must be done by another module. Yearly disturbance events are read in during the `annual` event so the only requirement is that the disturbances are summarized & available in this table by the time that event runs for the year. In other words: you could read the disturbances all at once in the beginning, or you could generate these values on a yearly basis during the simulation run.

2. `disturbanceMeta` replaces `mySpuDmids`

- `mySpuDmids` gets a rename that better represents its close association with the `disturbanceEvents` table
- The `rasterID` is renamed to `eventID`
- A new (optional) column named `priority` has been added where the user can set which events gets precedence over others in the case where multiple events happen in one pixel. For example, if you have:

```
## disturbanceEvents
| pixelIndex| year| eventID|
|----------:|----:|-------:|
|          1| 2025|       4|
|          1| 2025|       8|

## disturbanceMeta (not all columns shown here)
| eventID| priority|name |
|-------:|--------:|:----|
|       4|        2|bugs |
|       8|        1|fire |
```

for pixel 1 in the year 2025, eventID 8 will be chosen for the pixel (name = "fire", priority = 1)


DominiqueCaron:
Yes, I think that would make the link with `LandR`. As you said, `LandRCBM_split3pools` could update `disturbanceEvents` yearly based on the fire/harvest/bug module.

DominiqueCaron (approved):

cboisvenue:
Yes, this all makes sense to me and it is in line with our discussions about have CBM_core be a-spatial.

cboisvenue (approved):

PR #46, opened 2025-03-31
module dependency update

suz-estella:
Did a clean up of the module required package list. The GHA should pass successfully showing that there's not any packages missing.

Removed: 
- magrittr: not used in module. There's a commented out use of a magrittr pipe (`%>%`) so I replaced it with the R pipe (`|>`) just in case.
- RSQLite: not used by this module (used by others e.g. CBM_defaults: it is listed there).
- SpaDES.core: doesn't need to be listed as it is the package running the module (will always be available).
- ggplot2, quickplot, and cowplot removed: these are dependencies of CBMutils and not called directly in the module.

### DIFF
--- a/CBM_core.R
+++ b/CBM_core.R
@@ -10,9 +10,8 @@ defineModule(sim, list(
   citation = list("citation.bib"),
   documentation = list("README.txt", "CBM_core.Rmd"),
   reqdPkgs = list(
-    "data.table", "ggplot2", "quickPlot", "magrittr", "terra", "RSQLite", "cowplot",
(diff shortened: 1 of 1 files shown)


eliotmcintire:
> SpaDES.core: doesn't need to be listed as it is the package running the module (will always be available).

One reason to keep `SpaDES.core` is if you have a minimum version requirement ... which is likely ... ?

suz-estella:
I had that thought as well. My guess is it's listed because it calls SpaDES.core::Plots.

Issue #3, opened 2020-10-07
need legend on the emissions plot


cboisvenue:
The `CBMutils::carbonOutPlot()` fnct plots emissions as one line per gas (CH4, CO, CO2), and the bars for products on the same graph with two y-axis. We need a legend for the lines. Can't get the legend to appear in ggplot2.

PR #48, opened 2025-04-03
Spinup function added; spinup moved to its own event

suz-estella:
PRing these changes as a somewhat "midway" point in my goal to A) have `CBM_core` create and mange pixel groups (mostly completed here) and B) make `CBM_core` an aspatial module. These changes do not require any updates in other modules so I think it's a good stopping point to review and see that our tests are passing. The rest will be easy to implement within this new structure.

The `spadesCBM` tests are passing using this branch.

## Changes

The main change here is the addition of the `Spinup` and `SpinupCohorts` functions. I thought it would be useful to capture the spinup process in a function to clarify exactly what information is needed for that process. Both of these functions explicitly list what columns are required and check that they are present. It also provides some easy defaults for columns that are rarely changes.

1. `SpinupCohorts`: helps to merge pixel, cohort, and growth curve information into a single table of cohort attributes with the required columns.
2. `Spinup`: Take the table of cohorts and the `growth_increments` and performs the spinup. It creates the pixel groups using the columns it needs and returns the results and a table acting as a key of `pixelIndex`:`pixelGroup` so you can map the results back to your input cohorts.

### Module updates

1. The spinup step has been moved to it's own event (`spinup`) and an `Init` function has been added for a place to do other initial work. The `sim$gcid_is_sw_hw` is created in `Init`.
2. The `level3DT` (pixel groups) object is now a module output instead of an input. It is created during the `postSpinup` event.
3. The `spatialDT` "ecozones" column is actually not used anywhere! In this update it is no longer required (it is OK if it is present, it is just ignored).
4. Parameters update: there are now 3 default values that can be assigned to each stand/cohort. All of these can instead be passed as a column in `spatialDT` where they can be set differently for each pixel.
- `historicDMtype` and `lastPassDMtype` inputs have been replaced with parameters `default_historical_disturbance_type` = 1, default_last_pass_disturbance_type` = 1. 
- `regenDelay` parameter renamed `default_delay` (defaults to 0)

@DominiqueCaron some highlights from reviewing the columns needed in the spinup:
- "gcids": this column is required but it functions only as a link to the `growth_increments` table. So: it wouldn't be hard to change the module to let the user set this column name to anything, however, maybe it's easiest to just generate our own unique ID and call it "gcids" every time.
- "ecozones": as mentioned above: no longer needed.
- All the columns required for the spinup are listed [here](https://github.com/suz-estella/CBM_core/blob/eeb20325b85f6bec5eb5eef0fb8823a5f394ac04/R/spinup.R#L10)

## What's next

This sets me up nicely to do these next. This will require some changes to our other modules (e.g. data prep).

1. Split `spatialDT` into `standDT` and `cohortDT`: a table with stand (pixel) attributes and a table with cohort attributes.
2. Final step to make module aspatial: Add an 'area' column to `standDT` so that we don't need the master raster to calculate the emissions totals.
3. Remove the step in data prep where `level3DT` is created (will have to make sure this works with `CBM_vol2biomass`). The `speciesPixelGroup` and `realAges` inputs will need to be changed in `CBM_core`.
5. If we like the `Spinup` function: add it to CBMutils with proper documentation and some tests

## Question about growth curve attributes

@cboisvenue @camillegiuliano @DominiqueCaron 

We should settle on how to best have `CBM_core` receive the `species_id` and `sw_hw` columns required for the spinup. Right now, we are expecting `growth_increments` table to have a `forest_type_id` column that we convert to the `sw_hw` column. The `species_id` is passed via the `speciesPixelGroup` input which will need to be updated. It would be good to find a more straightforward way to provide these.

I imagine one of the following is the way to go:

1. Require `growth_increments` to have the `sw_hw` (or `forest_type_id`?) and `species_id` columns. PROS: closely associated with the growth curves; CONS: creates a lot of extra rows for columns that are 1:1 with `gcids`.
2. Require `spatialDT`/`cohortDT` to have the `sw_hw` and `species_id` columns. PROS: easy to implement and understand. CONS: introduces redundant columns to a table we likely want to keep as minimal as possible.
3. Introduce a growth curve metadata input (like `gcMeta`) to the module with the columns `gcids`, `species_id`, `sw_hw` so that it can be joined in during the spinup. PROS: closely associated with the growth curves; one row for each gcid (RAM friendly). CONS: adds another input object to the module (perhaps OK with good metadata).


DominiqueCaron:
Cool!

I am modifying things on my side as well. I have a column in `cohortDT` and `growth_increments` (I called it growthIncrements, but whatever) called `gcIndex` that links the cohort to the correct growth curve. Would it be to simply change it to `gcids`?

For `species_id` and `sw_hw`, aren't they cohort-level attributes? I don't see why they could not just be columns in `cohortDT`. With `LandR`, many pixels will likely have a mix of hardwood and softwood species.

This makes me think about something. For the below ground biomass/DOM, does sw/hw matters @cboisvenue ? If we keep track of below ground biomass at the pixel/stand-level we need to decide what we define as a sw stand vs hw stand. Does having proportion of each possible?

And last thing: for now, I don't use "spatial_units", I always directly use ecozones and jurisdiction. How do you determine the "spatial_units" in CBM? Is there a metadata table to determine which spatial_units is ecozone X and jurisdiction Y?

cboisvenue:
Wow. Nice work Susan. You're in fire. Here are some things that come up as I read this:
- I like the Spinup functions. That would eventually bring the Python function for spinup into CBMutils...are you planning on moving all the Python-functions to CBMutils? What does that mean for our reticulate use and the Python environments?
- This is silly but, I associate any function that uses pascal case (LikeSpinup), as opposed to camel case (likeSpinup) with people  outside the SpaDES group. Can we make it "spinup" instead of "Spinup"? If you have a reason you want it that way, please keep it. If not, can we change it?
- ecozones are how some parameters in CBM (from the defaults) are associated with a record (for decay or turnover for example). They are also how the Boudewyn parameters are linked to Boudewyn equation parameters. The Boudewyn thing might not be important since those calculations are already all done by the time we get to CBM_core. But the link to CBM parameters...when I checked the `*.csv` files, they use `spatial_unit_id` (see [turnover](libcbm/resources/cbm_exn/turnover_parameters.csv)). maybe the spatial_unit is enough? I don't know. We have to check the Python functions for that (Note: if we check the step function (libcbmr::cbm_exn_step), we will know since the spinup function calls the step function). Once that is checked, and we know the link to the parameters is via `spatial_unit_id` (not ecozone), then let's remove it.
- naming something "stand" can be dangerous as a bunch of people have a different idea of what a forest stand is...but you say is it a pixel. As long as "standDT" is clearly defined as pixel-level, I think this will work (but let's make sure we make that clear in the code and in the documentation).
- `forest_type_id` comes from the `species.csv` table. Maybe we can make it the `sw_hw` flag when we read the species in in `CBM_defaults`? This would get rid of the `forest_type_id` label/column.
- the updates to the module look good. One thing to make clear, and maybe we want to change this, `lastPassDMtype` is there so users can put a clear-cut instead of a fire as a last pass disturbance before the pixel/stand is grown to its inventory age. Setting a hard coded default might make that a bit harder? Maybe it deserves a message? It would have to be clear in the documentation that this is the default and how to change it.
- `level3DT` was originally used in the spinup to reduce the number of records. But I see you do that on line 27 in [sinpup.R](R/spinup.R). We'll have to figure out how `CBM_vol2biomass` can do without `level3DT`.
- growth curve attributes: all your proposed options would work. I think I prefer number 3. But could go either way
- @DominiqueCaron: yes, `sw_hw` is important for DOM dynamics in the turnover process (snags and roots). The parameters are [here](libcbm/resources/cbm_exn/turnover_parameters.csv) for our runs. The `sw_hw` flag is not important for the [decay](libcbm/resources/cbm_exn/decay_parameters.csv), but that is something I can easily imagine a research wanting to change (meaning: different tree species have different proportions of lignin, cellulose, hemicellulose, and pectin which decay at different rates). So, as time goes by and turnover finishes, that flag (`sw_hw`) won't be useful anymore.

Let me know if something is not clear of if I missed anything. Thanks for the good work!

suz-estella:
Responding to @DominiqueCaron:

> I am modifying things on my side as well. I have a column in `cohortDT` and `growth_increments` (I called it growthIncrements, but whatever) called `gcIndex` that links the cohort to the correct growth curve. Would it be to simply change it to `gcids`?

That's right! `gcIndex` is a nice name if we ever decide to rename it.

> For `species_id` and `sw_hw`, aren't they cohort-level attributes? I don't see why they could not just be columns in `cohortDT`. With `LandR`, many pixels will likely have a mix of hardwood and softwood species

Definitely agree that these columns are not stand/pixel attributes and they would need to be set for every cohort. However: I'm thinking that they are really growth curve attributes (if I'm correct that every gcid is 1:1 to the `species_id` and `sw_hw` attributes). I'm debating if they should be kept somewhere other than the cohort table then joined in via `gcids` key for when they are needed to keep the cohort table as minimal as possible (options 1 or 3). I think having it in the cohorts table could be fine though (option 2).

> And last thing: for now, I don't use "spatial_units", I always directly use ecozones and jurisdiction. How do you determine the "spatial_units" in CBM? Is there a metadata table to determine which spatial_units is ecozone X and jurisdiction Y?

This has to be joined in via the CBM defaults DB `spatial_unit` table using the `admin_boundary_id` and `eco_boundary_id` keys. I could imagine a `CBMutils` function could be useful here since it's something that we need to do a lot.

This query brings in the ecozone and admin boundary names:
```
> dbPath <- "~/data/cbm_defaults_v1.2.8340.362.db"
> cbmDB <- DBI::dbConnect(RSQLite::dbDriver("SQLite"), dbname = dbPath)
> spatialUnits <- DBI::dbGetQuery(cbmDB, paste(
  "SELECT a.id AS spatial_unit_id,",
  "a.admin_boundary_id, b.name AS admin_name,",
  "a.eco_boundary_id, c.name AS eco_name",
  "FROM spatial_unit a",
  "LEFT JOIN admin_boundary_tr b ON a.admin_boundary_id = b.admin_boundary_id",
  "LEFT JOIN eco_boundary_tr c ON a.eco_boundary_id = c.eco_boundary_id",
  "WHERE b.locale_id = 1 AND c.locale_id = 1",
  "ORDER BY a.id"))

> head(spatialUnits, 5)
  spatial_unit_id admin_boundary_id           admin_name eco_boundary_id           eco_name
1               1                 1         Newfoundland              17 Boreal Shield East
2               3                 2             Labrador              16  Taiga Shield East
3               4                 2             Labrador              17 Boreal Shield East
4               5                 3          Nova Scotia               7  Atlantic Maritime
5               6                 4 Prince Edward Island               7  Atlantic Maritime
```

suz-estella:
@cboisvenue and responding to yours!

> This is silly but, I associate any function that uses pascal case (LikeSpinup), as opposed to camel case (likeSpinup) with people  outside the SpaDES group. Can we make it "spinup" instead of "Spinup"? If you have a reason you want it that way, please keep it. If not, can we change it?

This would clash with the `spinup` event function! Maybe I could rename it "cbmExnSpinup"?

> I like the Spinup functions. That would eventually bring the Python function for spinup into CBMutils...are you planning on moving all the Python-functions to CBMutils? What does that mean for our reticulate use and the Python environments?

I have been thinking about this! The cleanest ways to go I think are A) leave any `libcbmr` functions as `CBM_core` modules functions and keep `libcbmr` out of `CBMutils` or B) write `CBMutils` wrapper functions of all the `libcbmr` calls (like `Spinup`) and remove `libcbmr` as a module dependency.

All that would be left to do to achieve option B would be write a `libcbmr::cbm_exn_step` function akin to `Spinup` and put that in `CBMutils` too. One thing that might be nice about having them in `CBMutils` is that we could keep tests for them there.

On the same topic: we could consider having the `reticulate::virtualenv_create/ReticulateFindPython` code block moved to the `Init` event of `CBM_core`.

> ecozones are how some parameters in CBM (from the defaults) are associated with a record (for decay or turnover for example). They are also how the Boudewyn parameters are linked to Boudewyn equation parameters. The Boudewyn thing might not be important since those calculations are already all done by the time we get to CBM_core. But the link to CBM parameters...when I checked the `*.csv` files, they use `spatial_unit_id` (see [turnover](libcbm/resources/cbm_exn/turnover_parameters.csv)). maybe the spatial_unit is enough? I don't know. We have to check the Python functions for that (Note: if we check the step function (libcbmr::cbm_exn_step), we will know since the spinup function calls the step function). Once that is checked, and we know the link to the parameters is via `spatial_unit_id` (not ecozone), then let's remove it.

Aye aye. I am almost certain that by the time `CBM_core` is running the spinup and annual events the `ecozone` is no longer needed even though it may be used often in data prep. In the next PR I'll rewrite the tests to not provide the `ecozones` column to be 100% sure.

> naming something "stand" can be dangerous as a bunch of people have a different idea of what a forest stand is...but you say is it a pixel. As long as "standDT" is clearly defined as pixel-level, I think this will work (but let's make sure we make that clear in the code and in the documentation).

OK noted!! We can certainly go with "pixel" but I'm also happy to write some clear documentation about it. Perhaps we can sit on it and debate further in the next PR where the tables get redefined.

> `forest_type_id` comes from the `species.csv` table. Maybe we can make it the `sw_hw` flag when we read the species in in `CBM_defaults`? This would get rid of the `forest_type_id` label/column.

I agree that it makes sense to just expect a `sw_hw` column instead of the `forest_type_id` column.

> growth curve attributes: all your proposed options would work. I think I prefer number 3. But could go either way

I think I lean towards #3 as well. Another thing I can draft up, PR, and put up for discussion as well.

> the updates to the module look good. One thing to make clear, and maybe we want to change this, `lastPassDMtype` is there so users can put a clear-cut instead of a fire as a last pass disturbance before the pixel/stand is grown to its inventory age. Setting a hard coded default might make that a bit harder? Maybe it deserves a message? It would have to be clear in the documentation that this is the default and how to change it.

For sure! I do have in the metadata documentation that there's 2 ways to set these values to override the default (via parameter or via a column). I'll definitely add a message but if you think it's better to have the last pass DM type not have a default (force the user to set) we can do that.

> `level3DT` was originally used in the spinup to reduce the number of records. But I see you do that on line 27 in [sinpup.R](R/spinup.R). We'll have to figure out how `CBM_vol2biomass` can do without `level3DT`.

Definitely have my eye on that! It's probably the trickiest part but I'm sure it's possible, perhaps with some clever event timing.

camillegiuliano:
>  1.  Require growth_increments to have the sw_hw (or forest_type_id?) and species_id columns. PROS: closely associated with the growth curves; CONS: creates a lot of extra rows for columns that are 1:1 with gcids.
> 2. Require spatialDT/cohortDT to have the sw_hw and species_id columns. PROS: easy to implement and understand. CONS: introduces redundant columns to a table we likely want to keep as minimal as possible. 
> 3.  Introduce a growth curve metadata input (like gcMeta) to the module with the columns gcids, species_id, sw_hw so that it can be joined in during the spinup. PROS: closely associated with the growth curves; one row for each gcid (RAM friendly). CONS: adds another input object to the module (perhaps OK with good metadata).

I also think 3 might be the best option here. We could also do a mix of both where the model looks for a metadata file (option 1) and if it doesn't receive it as an input, builds a "default" with option 1.

> I agree that it makes sense to just expect a sw_hw column instead of the forest_type_id column.

Yep agreed, this is an easy change I can make to our `CBMspecies` object (which is the `species.csv` table) in CBM_defaults if we want this. 

> We'll have to figure out how CBM_vol2biomass can do without level3DT.

Currently in vol2biomass (both SK and RIA), `level3DT` is only used when building building the unique gcids with `curveID`, but there is a check there and if `level3DT` isn't present it looks like they are being created anyway. So the module should work fine without it, I think. 
The errors I'm hitting now with the RIA are related to `level3DT` using the "old" `gcids` which don't match with the ecozone-specific `gcids` from the vol2biomass outputs. Creating `level3DT` in core might also help solve this issue (or at least simplify the solution).

cboisvenue:
@suz-estella responding back:
> This would clash with the `spinup` event function! Maybe I could rename it "cbmExnSpinup"?

Ok. please name it "cbmExnSpinup"

> I have been thinking about this! The cleanest ways to go I think are A) leave any `libcbmr` functions as `CBM_core` modules functions and keep `libcbmr` out of `CBMutils` or B) write `CBMutils` wrapper functions of all the `libcbmr` calls (like `Spinup`) and remove `libcbmr` as a module dependency.
> 
> All that would be left to do to achieve option B would be write a `libcbmr::cbm_exn_step` function akin to `Spinup` and put that in `CBMutils` too. One thing that might be nice about having them in `CBMutils` is that we could keep tests for them there.
> 
> On the same topic: we could consider having the `reticulate::virtualenv_create/ReticulateFindPython` code block moved to the `Init` event of `CBM_core`.

Again, I think either way could work. I really like moving the `reticulate::virtualenv_create/ReticulateFindPython` code block to the `CBM_core` `Init` event. Maybe all Python in `CBM_core` is better? You can decide on that as I don't really see any unwanted implication of either.

> Aye aye. I am almost certain that by the time `CBM_core` is running the spinup and annual events the `ecozone` is no longer needed even though it may be used often in data prep. In the next PR I'll rewrite the tests to not provide the `ecozones` column to be 100% sure.

Great. Thank you.

> OK noted!! We can certainly go with "pixel" but I'm also happy to write some clear documentation about it. Perhaps we can sit on it and debate further in the next PR where the tables get redefined.

I think `standDT` is ok, just keep the documentation clear.

> I think I lean towards #3 as well. Another thing I can draft up, PR, and put up for discussion as well.

We're all leaning towards #3. So let's do that.
 
> For sure! I do have in the metadata documentation that there's 2 ways to set these values to override the default (via parameter or via a column). I'll definitely add a message but if you think it's better to have the last pass DM type not have a default (force the user to set) we can do that.

I don't think we need to force the user. We just have to make it clear that they can change it. 

> Definitely have my eye on that! It's probably the trickiest part but I'm sure it's possible, perhaps with some clever event timing.

We (as @camillegiuliano pointed out) had another option just in case there was no `level3DT`, but it has never been tested. That is a poor name btw...the `level3DT` only means something to me. We could give it a better name that means something.

DominiqueCaron:
About the growth curve attributes. When `LandR` will provide the annual increments, the growth curve id will change each year? 

Actually, it will not be growth curve _per se_, `LandR` will simply provide how much the biomass of cohort _i_ in pixel _j_ change. What I was envisioning was that during the annual event `gcids` and `cohortID` will be the same. 

So first, is that a problem? 

Second, in that case, `gcMeta` becomes a bit awkward since would need to change yearly and it's already 1 line per gcids in `growth_increments`. I could just add `species_id` (that's the canfi species code, right?) and `sw_hw` attributes to the `growth_increments` object. Then. in `CBM_core`, it would retrieve the attributes from `gcMeta`, only if they are not in `growth_increments`. Does that make sense?

cboisvenue:
> Responding to @DominiqueCaron:
> 
> > I am modifying things on my side as well. I have a column in `cohortDT` and `growth_increments` (I called it growthIncrements, but whatever) called `gcIndex` that links the cohort to the correct growth curve. Would it be to simply change it to `gcids`?
> 
@suz-estella 
Your query (below) needs to be incorporated in `CBM_defaults`, it can replace the external link to `cbmAdmin`. 
@camillegiuliano, can `CBM_vol2biomass` run without `cbmAdmin`? or maybe we keep the link to the google drive version of `cbmAdmin` if the `CBM_vol2biomass` runs independently and use the one that comes out of `CBM_defaults` (or will come out of - can you add that too?)  when modules are run together?

> This query brings in the ecozone and admin boundary names:
> 
> ```
> > dbPath <- "~/data/cbm_defaults_v1.2.8340.362.db"
> > cbmDB <- DBI::dbConnect(RSQLite::dbDriver("SQLite"), dbname = dbPath)
> > spatialUnits <- DBI::dbGetQuery(cbmDB, paste(
>   "SELECT a.id AS spatial_unit_id,",
>   "a.admin_boundary_id, b.name AS admin_name,",
>   "a.eco_boundary_id, c.name AS eco_name",
>   "FROM spatial_unit a",
>   "LEFT JOIN admin_boundary_tr b ON a.admin_boundary_id = b.admin_boundary_id",
>   "LEFT JOIN eco_boundary_tr c ON a.eco_boundary_id = c.eco_boundary_id",
>   "WHERE b.locale_id = 1 AND c.locale_id = 1",
>   "ORDER BY a.id"))
> 
> > head(spatialUnits, 5)
>   spatial_unit_id admin_boundary_id           admin_name eco_boundary_id           eco_name
> 1               1                 1         Newfoundland              17 Boreal Shield East
> 2               3                 2             Labrador              16  Taiga Shield East
> 3               4                 2             Labrador              17 Boreal Shield East
> 4               5                 3          Nova Scotia               7  Atlantic Maritime
> 5               6                 4 Prince Edward Island               7  Atlantic Maritime
> ```

camillegiuliano:
> can CBM_vol2biomass run without cbmAdmin?

Currently not, we subset `cbmAdmin` to the study area's spatial unit id and ecozone to further subset the Boudewyn tables into the smaller tables (`stableX`) using `EcoBoundaryID` (ecozone) and `abreviation` (province name abreviations).

cboisvenue:
> About the growth curve attributes. When `LandR` will provide the annual increments, the growth curve id will change each year?
> 
> Actually, it will not be growth curve _per se_, `LandR` will simply provide how much the biomass of cohort _i_ in pixel _j_ change. What I was envisioning was that during the annual event `gcids` and `cohortID` will be the same.
> 
> So first, is that a problem?

When we link to LandRCBM, the yearly growth increment will come from LandRCBM. Right now in spadesCBM, it comes from looking up the the growth curve ("gcids") at the specific age that records is at. 

This is how I understand the LandRCBM, please correct this thinking if it is wrong: 
1) during the spinup event, there will be one gcid per pixel but there can be multiple cohorts (each with their growth curve) in one pixel and at the end of the spinup, each cohort has "full" carbon pools. Pools are tracked by cohort but the grow-and-burn cycle is tracked by pixel. Right? I am unclear how below ground or DOM pools at the pixel-level are going to be split back in the cohort vector. I think I am missing something here (forgot to take a photo of Dominique's explanation on my office white board).
2) on an annual basis, all cohorts will have one biomass increment that is a product of the LandR simulation. So each cohort needs its increment, so it will have its unique gcid. 

Note that most
> 
> Second, in that case, `gcMeta` becomes a bit awkward since would need to change yearly and it's already 1 line per gcids in `growth_increments`. I could just add `species_id` (that's the canfi species code, right?) and `sw_hw` attributes to the `growth_increments` object. Then. in `CBM_core`, it would retrieve the attributes from `gcMeta`, only if they are not in `growth_increments`. Does that make sense?

I think this makes sense. When CBM will be driven by LandR, the record or "unit-on-which" growth happens changes from the stand (which in spadesCBM is equivalent to the pixel) to the cohort. We will have more record/vectors.

DominiqueCaron:
> This is how I understand the LandRCBM, please correct this thinking if it is wrong:
> 
> 1. during the spinup event, there will be one gcid per pixel but there can be multiple cohorts (each with their growth curve) in one pixel and at the end of the spinup, each cohort has "full" carbon pools. Pools are tracked by cohort but the grow-and-burn cycle is tracked by pixel. Right?

Actually, after our discussion, I changed the terminology in LandRCBM slightly to be less confusing with CBM. During the spinup each pixel has a `yieldTableIndex` which points to the growth curves of each cohort within that pixel. These cohort-specific growth curves are named `gcids`. So, in LandRCBM, a "yield table" would be the "growth curve" of each species within the area. So during the spinup, it is similar to how CBM works now, the difference is that it will track carbon per cohort instead of stand (each cohort as a `gcids`. So the rest of your interpretation is correct.

> 2. on an annual basis, all cohorts will have one biomass increment that is a product of the LandR simulation. So each cohort needs its increment, so it will have its unique gcid.

Right!

suz-estella:
Pushed a few changes here:
1. Renamed the `Spinup` function `cbmExnSpinup` and `SpinupCohorts` function `cbmExnSpinupCohorts`
2. Removed the "ecozones" column from `spatialDT` in the test data so now our tests demonstrate that the `ecozones` column is not needed in `spatialDT`
3. Added messages about the use of default parameters in the spinup (e.g last_pass_disturbance_type)

I think there's just one thing that needs to be resolved before merging this particular PR: can `CBM_vol2biomass` live without `level3DT`? `CBM_dataPrep_SK` is still creating `level3DT` for now so merging this won't be an issue but I think it makes sense to have a plan about how to address this.

Looking at `CBM_vol2biomass`: it seems like it is working fine with `spatialDT` directly, and I suspect it might be most efficient that way -- it only needs to work with unique combos of `gcids`, `spatial_unit_id`, `ecozone` -- not the `ages` column. In simpler words: it needs less "pixel groups" than `CBM_core` so perhaps it's OK making its own. I suspect the other use of `level3DT` could be replaced by either using `spatialDT` or `gcMeta` (or something). @camillegiuliano what do you think?

Otherwise: this has been a great convo! I tried to identify all the different topics/todo's left to make new task drafts out of to address one by one. I'll link them all back to this convo.

- Update `CBM_dataPrep_SK` to no longer create `level3DT` and `CBM_vol2biomass` to no longer use it
- Split `spatialDT` into `standDT` and `cohortDT` (and maybe rename `level3DT` once and for all while I'm at it)
- Consider adding a 'gcMeta' input table with the `species_id` and `sw_hw` for each growth curve (@DominiqueCaron @cboisvenue I can see there is more to discuss there -- maybe gcMeta could be updated yearly?)
- Add the `area` column to `standDT` so the module no longer has any spatial objects left in it (will need to discuss how plotting fits into this at some point)
- Review the possibility of replacing `cbmAdmin` with a direct query to `spatial_units`
- Move the Python set up to the `Init` event; add `cbmExnSpinup` to CBMutils; consider creating a new function `cbmExnStep` so that all the `libcbmr` processes are wrapped in functions

DominiqueCaron:
Awesome @suz-estella ! I think all that will make linking LandR way easier! some final thoughts:

> Split spatialDT into standDT and cohortDT (and maybe rename level3DT once and for all while I'm at it)

With `LandR`, I believe that `cohortDT` will already be created in `Biomass_split3pools`. I wonder if that could be a simple if(exists("sim$cohortDT")). Is the pixel-level attribute output of `Biomass_split3pools` be called `standDT` or `pixelDT`?

> Consider adding a 'gcMeta' input table with the species_id and sw_hw for each growth curve (@DominiqueCaron @cboisvenue I can see there is more to discuss there -- maybe gcMeta could be updated yearly?)

Yes, I guess gcMeta could be updated yearly!

suz-estella:
> With `LandR`, I believe that `cohortDT` will already be created in `Biomass_split3pools`. I wonder if that could be a simple if(exists("sim$cohortDT")). Is the pixel-level attribute output of `Biomass_split3pools` be called `standDT` or `pixelDT`?

I'm thinking 'standDT'! I'll be sure to make you aware of the PR once it's all together. I tried to prep the `cbmExnSpinupCohorts` function with this in mind so I'm thinking the column lists for each table will look close to this: https://github.com/suz-estella/CBM_core/blob/fc19db15631179dc57aea24faa97602187ea101d/R/spinup.R#L85

PR #50, opened 2025-04-07
allows increments to trail +100 if pixelGroup age exceeds that of the growth curves

camillegiuliano:
checks if pixelGroup ages exceed the max age of the growth curves, if yes, trails the final value an extra 100 years.

### DIFF
--- a/CBM_core.R
+++ b/CBM_core.R
@@ -271,7 +271,6 @@ doEvent.CBM_core <- function(sim, eventTime, eventType, debug = FALSE) {
 }
 
 Init <- function(sim){
-
   # Determine if growth curves are for SW or HW
   sim$gcid_is_sw_hw <- unique(sim$growth_increments[, .(gcids, is_sw = forest_type_id == 1)])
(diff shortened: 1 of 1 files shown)


cboisvenue:
@camillegiuliano GHA are failing. Is that still because of a problem with the ecozone.zip?

camillegiuliano:
Ah! Looks like SK doesn't like the tweak I made to get the growth_increment object to build properly in the RIA, let me go investigate why it's failing

camillegiuliano:
Ok this should pass, we'll see. 
It was because `sim$speciesPixelGroup` has more columns in the RIA than SK, including a gcid column. Merging with both `pixelGroup` and `gcids` was fine in the RIA and solved the double gcid column issue, but was impossible in SK since the column didn't exist.
This change instead creates a version of `speciesPixelGroup` outside of the simlist that only has the columns it needs for this merge. This solves the double `gcids` column in the RIA and doesn't change anything for SK

PR #51, opened 2025-04-07
'gcMeta' input object replaces 'speciesPixelGroup' input and 'gcid_is_sw_hw' output

suz-estella:
As discussed in #48 (@cboisvenue, @DominiqueCaron, @camillegiuliano)

The module has a new input called `gcMeta` table with the columns `gcids`, `species_id`, and `sw_hw`.

- 'speciesPixelGroup' (which provided the `species_id` for the spinup) is no longer a required input
- `gcid_is_sw_hw` output is not longer created because that information is already readily available in `sim$gcMeta`.

marked as a draft until resolved: https://github.com/PredictiveEcology/CBM_vol2biomass/pull/20


suz-estella:
@cboisvenue a note: the integration tests are going to fail here because it will only work with this version of `CBM_dataPrep_SK`: https://github.com/PredictiveEcology/CBM_dataPrep_SK/pull/49

The tests pass locally if I manually set the branch location there. I'll re-run everything after merging.

PR #53, opened 2025-04-15
Python virtual environment created in Init event

suz-estella:
Python virtual environment `r-spadesCBM` created in the module Init event. It no longer needs to be initiated in the `setupProject` call.


cboisvenue:
@camillegiuliano all our globals will need to be updated to remove the Python/reticulate portion.

camillegiuliano:
On it, I'll change the spadesCBM@main ones as well, since I'll be merging @development into @main today anyway!

camillegiuliano:
RIA global is changed, didn't touch the SK ones since Susan already PR'ed those :)

PR #54, opened 2025-04-17
Bug fix: growth increment setting for new pixel groups

suz-estella:
This fix ensures that growth increments are assigned to the new set of pixel groups by joining to `sim$spinup_input$increments` via the current age and pixel group's original (spinup, before disturbances) group ID. I was able to observe that these were being assigned incorrectly in this section with our SK small test data. The validation data has been updated to reflect the results when the increments are assigned correctly.

I also added a warning where the growth increments need to be extended due to the cohort ages exceeding what is available in the table. This was first observed when running RIA in 2015 where there is an age of 322 but the increments only projected to 250 (may have been 300). @cboisvenue would this be better as a message instead of a warning? The text is: `Cohort ages exceed growth increment ages. Increments for the greatest available age have been applied to older cohorts.`


cboisvenue:
@suz-estella I think a warning is fine. This will happen often especially as we use landR to drive CBM.
Nice work here!

PR #59, opened 2025-04-23
spatialDT now uses the 'area' column to calculate emissions totals

suz-estella:
Follows https://github.com/PredictiveEcology/CBM_dataPrep_SK/pull/53

The pixel areas within each pixel group are summed together to get a total "group area" which is then multiplied by the emissions. This allows for the pixel areas to differ in size if need be. This could potentially be relevant if we wanted to run the module with a non-projected (lat-long) projection or if we want to run the module by merging input tables from multiple study areas. Maybe unlikely but the option is there!

the `masterRaster` is no longer required for any part of the module's essential events, but it can be used in plotting if it is available in the `simList`. For this reason it hasn't been removed from the module input list entirely.

### DIFF
--- a/CBM_core.R
+++ b/CBM_core.R
@@ -45,13 +45,11 @@ defineModule(sim, list(
   ),
   inputObjects = bindrows(
     expectsInput(
-      objectName = "masterRaster", objectClass = "raster",
-      desc = "Raster has NAs where there are no species and the pixel `groupID` where the pixels were simulated. It is used to map results"),
-    expectsInput(
       objectName = "spatialDT", objectClass = "data.table", sourceURL = NA,
       desc = "Table of pixel attributes",
       columns = c(
         pixelIndex      = "Stand ID",
+        area            = "Stand area in meters",
         spatial_unit_id = "CBM-CFS3 spatial unit ID",
         gcids           = "Growth curve ID",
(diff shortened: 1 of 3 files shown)


cboisvenue:
@suz-estella looks like this is failing windows. I look at it and it is in the test-multiModule_SK-small_1998-2000.R test. The error message indicates that the annual(sim) function failed because the spatialDT object is missing the "area" column, which you need to calculate emissions and product totals.

suz-estella:
> @suz-estella looks like this is failing windows. I look at it and it is in the test-multiModule_SK-small_1998-2000.R test. The error message indicates that the annual(sim) function failed because the spatialDT object is missing the "area" column, which you need to calculate emissions and product totals.

that's exactly what I'm hoping to see! the area column isn't created by `CBM_dataPrep_SK` until this PR is merged: https://github.com/PredictiveEcology/CBM_dataPrep_SK/pull/53. It's a good sign that the module tests are passing where I have altered the input to provide the area column.

suz-estella:
@cboisvenue all failed with the same error (no area column). we could either merge the `CBM_dataPrep_SK` update in and re-run these or merge both at the same time and I can keep an eye that the GHAs triggered by the merge all pass.

PR #60, opened 2025-04-24
spatialDT split into standDT and cohortDT

suz-estella:
As discussed in this PR: https://github.com/PredictiveEcology/CBM_core/pull/48

Marked as a draft until I PR a corresponding update to `CBM_dataPrep_SK`. The GHA will fail at the integration test step but the module tests should pass (I will confirm that I see this result in a comment).

## Description

`spatialDT` has been split into `cohortDT` and `standDT` to allow for there to be multiple cohorts per stand (per pixel). The goal was to split these tables into which attributes describe the stand's physical state and which attributes describe each individual cohort.

`cohortDT` columns: 
- `pixelIndex`
- `area`
- `spatial_unit_id`
- `historical_disturbance_type` (optional)
- `last_pass_disturbance_type` (optional)

`standDT` columns: 
- `cohortID`
- `pixelIndex`
- `gcids`
- `ages`
- `ageSpinup` (optional)
- `delay` (optional)

The annual event now works with `cohortGroupIDs` instead of `pixelGroups`. `pixelGroupC` has been renamed `cohortGroups` and `pixelKeep` has been renamed `cohortGroupKeep`.

## Remaining questions

@cboisvenue @DominiqueCaron this brings up a question that I haven't attempted to resolve in this PR: How do we handle calculating the emissions and products totals when there are multiple cohorts per stand?

Currently, the `spinup` and `annual` events work with the pixels as if they are 1m x 1m resolution, and later the `cbm_vars$flux` and `cbm_vars$pools$Products` values are multiplied by the true area of each pixel to calculate the totals. Is this still valid when there are multiple cohorts per stand? Or does the presence of the other cohorts affect these values in a way not reflected in the growth curves/increments?

@cboisvenue I imagine it makes sense to incorporate the standard CBM-CFS3 method here, but linking you in @DominiqueCaron as I'm sure it is relevant to your work even if LandR takes another approach.

To be safe: in this PR, the emissions totals are not calculated unless these is 1 cohort per stand, with a warning that the module is not ready to handle that scenario yet. So: we have the choice to deal with this now or later.


cboisvenue:
@suz-estella is it possible that you have the 'cohortDT' and 'standDT' columns reversed in this description? The other way around makes more sense to me.

I am not sure I understand your 1mX1m behaviour, In spadesCBM runs, increments are in tonnes of biomass per ha. So the size of the pixel will help us calculate the emmissions and products. @DominiqueCaron I think this is the case for LandR increments that come out of `LandRCBM_split3pools`, correct?

The "influence" of one cohort on the other is captured on the LandR side, on the CBMLandR connection, the increments are per cohort and that already takes into account that specific cohort environment. To get the total flux or products from a pixel, you would add all the affected cohorts, which gives you a total per ha, and multiply by the size of the pixel. Is this making sense? We could meet to clarify as I am unsure if I am answering your question.

suz-estella:
@cboisvenue: I did switch the table names by mistake! edited to fix.

Regarding pixel size -- that makes sense -- it's 1ha x 1ha cells. in the end, we divide the area in meters by 10,000 to calculate the totals.

For LandR that makes sense to me, I think. In summary: for LandR, the emissions totals will indeed be correct if we multiply the fluxes by the pixel area (in ha) regardless of how many cohorts are in a single pixel/stand. So that's easy. However: if we are trying to run a standard CBM-CFS3 simulation using `CBM_vol2biomass` where we have multiple cohorts/pixel: is this also the case?

DominiqueCaron:
Yes the output of `LandRCBM_split3pools` is also in tonnes of biomass per ha.

I also realized I added commits on accident....? I thought I was on my fork, I am confused on how that'd happen. Not sure we want those change in that PR.

suz-estella:
> Yes the output of `LandRCBM_split3pools` is also in tonnes of biomass per ha.
> 
> I also realized I added commits on accident....? I thought I was on my fork, I am confused on how that'd happen. Not sure we want those change in that PR.

@DominiqueCaron it looks like since your commit is older and also affects the `spinup` event the merge chose to override it with my commit, effectively leaving this branch the same. So, all good!

cboisvenue:
@suz-estella can we reverse Dominique's commits until we are sure we want them?

> For LandR that makes sense to me, I think. In summary: for LandR, the emissions totals will indeed be correct if we multiply the fluxes by the pixel area (in ha) regardless of how many cohorts are in a single pixel/stand. So that's easy. However: if we are trying to run a standard CBM-CFS3 simulation using `CBM_vol2biomass` where we have multiple cohorts/pixel: is this also the case?

hum. Not sure I understand your question: when we run standard spadesCBM simulations, the increments are in tonnes of carbon per ha and this comes from `CBM_vol2biomass` (note that I had written "tonnes of biomass per ha" above - it is now corrected) and our "stands" are really our pixels. So, we would still add up all the per ha values in the pixel and multiply by the size of the pixel (so 0.09 ha in the case of 30m pixels). What I am missing?

@DominiqueCaron are the values from `LandRCBM_split3pools` really in tonnes biomass per ha or are they in tonnes of carbon per ha? The current default conversion in standard spadesCBM simulations is to multiple by 0.5 (this value comes from [this table](libcbm/resources/cbm_exn/root_parameters.csv) but can be modified by user to match the wood density of specific species).

DominiqueCaron:
For the emissions and products totals, I am not sure to understand the problem either. It seems like whether it is LandR or standard CBM simulation, it shouldn't change. The totals would be the sum of everything in that pixel times the size of the pixel.

So on this line, we divide tonnes/ha of biomass by 0.5, so that makes it tonnes/ha of carbon, right? https://github.com/PredictiveEcology/CBMutils/blob/bd30825d38c1d8abe7d0d002d494745e19b42ec5/R/Boudewyn_cumPoolsCreateAGB.R#L61

suz-estella:
> @suz-estella can we reverse Dominique's commits until we are sure we want them?

The code is A-OK (my work took precedence) If we want to keep a clean history with this merge we could consider doing a "Squash and merge" instead of a regular "Merge".

> hum. Not sure I understand your question: when we run standard spadesCBM simulations, the increments are in tonnes of carbon per ha and this comes from `CBM_vol2biomass` (note that I had written "tonnes of biomass per ha" above - it is now corrected) and our "stands" are really our pixels. So, we would still add up all the per ha values in the pixel and multiply by the size of the pixel (so 0.09 ha in the case of 30m pixels). What I am missing?

This all checks out with what I understand about our current spadesCBM runs where we have 1 cohort per pixel. And yes: I don't see that we need to change anything here to make this work with our current projects (spadesCBM with 1 cohort per pixel and LandR).

Part of my question my just wondering whether or not this will be valid for LandR run, so that's great (and easy) that it is.

Although this is not something we are doing at the moment, we spoke briefly about how CBM-CFS3 (without LandR) can potentially be run with multiple cohorts/pixel, which is part of why this table split is generally useful. I want to be certain that `CBM_core` doesn't need to be doing anything special to account for this, say in the case that we tried this with `CBM_vol2biomass`. I believe this convo has answered my question: I'm getting the impression that regardless of which module produces the increments (`CBM_vol2biomass`, LandR, etc) it is their job to have already represented the cohort dynamics across the pixel area, so multiplying by the stand/pixel area will always produce the correct emissions totals.

If that's all correct: I'll simply revert the warning about this and we're good to go for having the totals always calculated.

cboisvenue:
@suz-estella you are correct. Please proceed!

suz-estella:
@cboisvenue good to go to be merged alongside: https://github.com/PredictiveEcology/CBM_dataPrep_SK/pull/54

The tests are all passing up until the integration test with the CBM_dataPrep_SK development branch but I'm able to run it successfully locally using the `suz-masterRaster` branches in these PRs.

@camillegiuliano I haven't touched the plotting section of this module but it is currently out of commission with this PR!

PR #64, opened 2025-05-07
annual event: growth incremenets now assigned directly from input sim$growth_increments

suz-estella:
A suggested simplification as first discussed in our meeting yesterday. I've noticed that the `sim$growth_increments` object is nearly identical to the `sim$spinupInput$increments` table except for two small differences:

1. `sim$growth_increments` has the original `gcids` column; `sim$spinupInput` drops the `gcids` column and instead has a `row_idx` column that matches with the cohort group IDs at the time of the spinup.
2. `sim$spinupInput$increments` does not have any rows for when age == 0

You can see where we create this in the cbmExnSpinup function: it's an object we have been making rather than an object that is returned by Python: [see here](https://github.com/PredictiveEcology/CBM_core/blob/f19b4fe7581ec895f120dd014633d3558b12a59f/R/spinup.R#L43) 

**Current behaviour**: In the annual event, cohort groups are assigned increments for `libcbmr::cbm_exn_step` from the `sim$spinupInput$increments` table via the current cohort `age` and the group's original (before disturbances) `spinup` group ID from `cohortGroupKeep`.

**Updated behaviour**: In the annual event, cohort groups are assigned increments for `libcbmr::cbm_exn_step` from the `sim$growth_increments` table via the current cohort `age` and the group's `gcid`.

The tests show that the results are exactly the same; the increments that are assigned are identical. It's just a more direct join. @cboisvenue is there some reason I'm not seeing as to why this was done in the first place and/or why we shouldn't get the increments directly from `sim$growth_increments` each year? How about when age == 0? In our SK example, this would mean that `merch_inc`, `foliage_inc`, and `other_inc` would be set to 0 when age == 0. Is that right?

Also up for discussion: I took this a step further and removed the `spinupInput` object from the module outputs entirely. As it is now, it's a list of only one item, the input "increments" table (was there maybe more assigned to this list at some point?). It is only ever used in the annual event instead of the `sim$growth_increments` object.

### DIFF
--- a/CBM_core.R
+++ b/CBM_core.R
@@ -125,9 +125,6 @@ defineModule(sim, list(
   ),
   outputObjects = bindrows(
     createsOutput(
-      objectName = "spinupInput", objectClass = "data.table",
-      desc = "Spinup inputs"),
-    createsOutput(
       objectName = "spinupResult", objectClass = "data.frame",
       desc = "Spinup results"),
     createsOutput(
@@ -332,8 +329,7 @@ spinup <- function(sim) {
     gcIndex    = "gcids"
   ) |> Cache()
 
-  # Save spinup input and output
-  sim$spinupInput  <- spinupOut["increments"]
+  # Save spinup output
   sim$spinupResult <- spinupOut$output$pools
 
   # Save cohort group key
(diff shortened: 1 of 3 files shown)


cboisvenue (approved):

camillegiuliano (approved):

DominiqueCaron (approved):
Looks good to me!

PR #65, opened 2025-05-15
disturbanceMeta update

suz-estella:
Another simplification to CBM_core.

This takes advantage of the fact that the annual Python process does not require anything other than the `disturbance_type_id` for each disturbed cohort. `disturbanceMeta` now only requires that each event has a `disturbance_type_id` set for it. The `disturbanceMeta$disturbance_matrix_id` column has been removed as a required input, as well as the `spatial_unit_id` and `sw_hw` columns that are there to identify each unique `disturbance_matrix_id`.

Currently, `libcbmr::cbm_exn_step` does not accept the `disturbance_matrix_id` as something we can control. The EXN `disturband_matrix_association` CSV is read in when [`libcbmr::cbm_exn_get_default_parameters`](https://github.com/PredictiveEcology/CBM_core/blob/39f31bcfb2da2b731210dbadf27dae81c8b9a56e/CBM_core.R#L606) is called and then this table is passed to `libcbmr::cbm_exn_step` along with `cbm_vars`. I assume somewhere in the process the `disturbance_matrix_ids` are automatically pulled by matching with the `cbm_vars` columns `cbm_vars$parameters$disturbance_type` (set using `disturbanceEvents` and `disturbanceMeta`), `cbm_vars$state$spatial_unit_id` (set from `standDT`), and `cbm_vars$state$sw_hw` (set from `gcMeta`).

If we want to have more control over the disturbance matrices in the future we'd have to find a way to make changes to this part of the process, perhaps by altering the `mod$libcbm_default_model_config` object. While I think this would be a worthwhile endeavour for research purposes I think it's maybe safest to remove these columns as listed inputs for now as it may imply that the module is already able to do this.

The reduces the workload of our data prep modules to prepare `disturbanceMeta` for CBM_core because there no longer needs to be a unique row for every `spatial_unit_id` and `disturbance_type_id`, and we no longer need to match it with the `disturbance_matrix_association` CSV ahead of time. I can PR following updates.
 
The following columns have been set as optional and their description left in the metadata in case they might be useful later, though they are not currently used (yet, perhaps):
```
  name        = "Optional. Disturbance name",
  description = "Optional. Disturbance description",
  wholeStand  = "Optional. Specifies if the whole stand is disturbed (1 = TRUE; 0 = FALSE)"
```

The tests are producing the same numbers; the only change is that there are now less pixel groups in the SK example because events 3 and 5 (20% mortality) have the same disturbance_type_id and are therefore treated as if they are the same.


suz-estella:
For some extra info: here's where libcbmr reads in disturbance_matrix_association.csv with `cbm_exn_get_default_parameters`: https://github.com/PredictiveEcology/libcbmr/blob/e70426361500824a580f9ecdc8cdb1e691144f5a/R/cbm_exn.r#L34

cboisvenue:
Nice work @suz-estella.
Here are a few comments:
- as you point out, to identify the right disturbance matrix (DM), the python functions for spinup (`libcbmr::cbm_exn_spinup()`) and for annual step (`libcbmr::cbm_exn_step()`) internally use `disturbance_type_id`, the `sw-hw` flag and an `spu`. I want the use to be able to know what DM they are using, and by that I mean the proportions that are used and the name of the DM. By removing this info from `disturbanceMeta` are we removing the power of the user to look at the disturbances being applied?
- Can Figure 1 in the CBM_core chapter of the manual still be built? @camillegiuliano 
- I don't have a good "view" of what is going on in the Python function. Like you said above we "assume somewhere in the process the disturbance_matrix_ids are automatically pulled by matching with the `cbm_vars` columns `cbm_vars$parameters$disturbance_type` (set using `disturbanceEvents` and `disturbanceMeta`), `cbm_vars$state$spatial_unit_id` (set from `standDT`), and `cbm_vars$state$sw_hw` (set from `gcMeta`)". So, I want to be able to show the match that we are assuming is being done. DMs are important to almost all users and are usually the first thing they want to see if they dig into anything.

suz-estella:
@cboisvenue why don't I see if there's a way that I could have the matrix IDs (and perhaps other useful info about the disturbances) saved or returned somewhere in the module outputs instead?

As it stands - the input `disturbance_matrix_id` is an unreliable source of info for reviewing what disturbances have been applied. It could contain anything. Our previous "EXN-mixup" debacle is a good example of this: previously, we were passing matrix IDs from the defaults DB (non-EXN) to CBM_core, but (thankfully) these were being disregarded: the EXN matrix IDs were still being used as brought in via the Python process. Reviewing the `disturbance_matrix_id` column sent to the module to understand what happened during the simulation would really only cause confusion to the user. Since it's listed as a required input column, I think it would be reasonable for someone to assume that they are indeed being used.

camillegiuliano:
Currently, the alluvial plot only uses `sim$cTransfers` which is from CBM_defaults. I don't see why we wouldn't be able to build it. 

Currently, you do need to know the `disturbance_matrix_id` of the disturbance to get the figure, but the table also has `disturbance_type_id` and `spatial_unit` so it shouldn't be a very hard to change to make.

cboisvenue:
> @cboisvenue why don't I see if there's a way that I could have the matrix IDs (and perhaps other useful info about the disturbances) saved or returned somewhere in the module outputs instead?
> 
> As it stands - the input `disturbance_matrix_id` is an unreliable source of info for reviewing what disturbances have been applied. It could contain anything. Our previous "EXN-mixup" debacle is a good example of this: previously, we were passing matrix IDs from the defaults DB (non-EXN) to CBM_core, but (thankfully) these were being disregarded: the EXN matrix IDs were still being used as brought in via the Python process. Reviewing the `disturbance_matrix_id` column sent to the module to understand what happened during the simulation would really only cause confusion to the user. Since it's listed as a required input column, I think it would be reasonable for someone to assume that they are indeed being used.

Ok @suz-estella I like this solution. We should not that if/when we or a user would like to change the actual proportions in the DMs, it would have to be done in the disturbance_matrix_association.csv.

cboisvenue (approved):

PR #66, opened 2025-05-28
Implement regeneration delay in the annual event

camillegiuliano:
This PR implements a regeneration delay to the annual event. Cohorts won't grow  (i.e. ages, merch_inc, foliage_inc, and other_inc will be 0'ed) until their `time_since_last_disturbance` value reaches the delay value. This delay is the same value that we set for the spinup (but if we want them to be independent I can change that). In this PR I have set the delay to 5 years. 

The way this works is by creating a TRUE/FALSE flag for each cohort, zero-ing age and growth (in the `cbm_vars$state` and `$parameters` tables) if the cohort is flagged as TRUE


suz-estella:
OK, interesting! In that case: I'm not sure if the delay is being used correctly in the spinup, then: should it always be 0?

cboisvenue:
In standard spadesCBM runs:
- the default delay should be 0 and the default should be used in the spinup.
- Yes Camille, we should be able to set the delay in the annual/step function to another value separately then the default/spinup.
- In standard spadesCBM, the behavior is that whatever curve was used before disturbance for that stand/cohort will be use to grow the new one.

In LandRCBM, I am not sure if we need any delays since each cohorts will just grow when LandR tells us it shows up (for new cohorts) or by the increment that LandR gives it based on all the competition interactions if simulates.

suz-estella:
As it stands now, the `default_delay` parameter and `delay` field are defined as representing the regeneration delay. So - I'm wondering if instead we might actually need to add a second parameter for the spinup delay if it represents another concept. 

Digging into [the docs](https://cat-cfs.github.io/libcbm_py/model_definition.html#libcbm.model.model_definition.model_variables.ModelVariables) to understand better!

suz-estella:
@cboisvenue to attempt to pose this as a simple question: do we need the ability to set the spinup delay as anything other than 0?

cboisvenue:
> @cboisvenue to attempt to pose this as a simple question: do we need the ability to set the spinup delay as anything other than 0?

@suz-estella I think we need to keep the ability to set it to something else then 0, but in our use-case, we will be setting it to 0. 

In standard spadesCBM, delays will be set to something else then 0 when we start to represent management that includes planting or simulations of delayed natural regeneration.

suz-estella:
Maybe it would make sense to have 2 parameters & 2 optional columns?:

```
    defineParameter(
      "default_delay_spinup", "integer", default = 0L, min = 0L, max = NA_integer_, desc = paste(
        "The default spinup delay.",
        "This can instead be set for each cohort with the spatialDT 'delaySpinup' column."
      )),
    defineParameter(
      "default_delay_regen", "integer", default = 0L, min = 0L, max = NA_integer_, desc = paste(
        "The default regeneration delay post disturbance.",
        "This can instead be set for each cohort with the spatialDT 'delayRegen' column."
      )),
```

cboisvenue:
This makes sense to me.

camillegiuliano:
I think it is good to merge for now. I set all both delays to 0 at the moment (i.e. exactly like in that code snippet Susan posted), but easy to change those values in the future with how it's currently set up

PR #68, opened 2025-05-30
regeneration delay can be set by parameter or by cohortDT column

suz-estella:
- Regeneration delay can be set via the `default_delay_regen` and/or the `cohortDT$delayRegen`  column (the parameter will fill NAs in the column if it exists)
- Test added for 2 cohorts: this checks that the delay can be applied both ways, and if so, the ages in the `cbmPools` output table will reflect this, and `NPP` will be lower for the cohort with the longer delay.
- Documentation updated

The main change here is the the `cbm_vars$state` table now holds the `delay` column. This is first added to the table in the `spinup` event and then kept through each `annual` event. It is ignored by Python and only used to implement the delay afterwards.

### DIFF
--- a/CBM_core.R
+++ b/CBM_core.R
@@ -23,22 +23,22 @@ defineModule(sim, list(
     defineParameter(
       "default_delay_spinup", "integer", default = 0L, min = 0L, max = NA_integer_, desc = paste(
         "The default spinup delay.",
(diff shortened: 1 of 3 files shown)


camillegiuliano (approved):

cboisvenue:
@camillegiuliano can we make sure how to handle delays is in the manual please?

PR #69, opened 2025-06-05
Products are now yearly instead of cumulative

camillegiuliano:
This tiny change allows the products column in the `emissionsProducts` object to represent values for the current year rather than cumulatively for the whole simulation.


cboisvenue:
Checks are failing... Do you know why?

camillegiuliano:
Ah, they are failing the test as the new Products values are not equal to those in the validation test values csv's. I'll do the runs associated with those tests to get the new values and update and then it should pass!

cboisvenue (approved):

Issue #72, opened 2025-06-19
CBM_core output review


suz-estella:
First discussed here: https://github.com/PredictiveEcology/CBM_core/pull/62#issuecomment-2855121706

The goal of this is to make the outputs of CBM_core more "complete" in that they contain everything required for reporting and plotting without the presence of any input or intermediate objects. This would make it possible to review the results long after a simulation is run without any missing context, and it would reduce the number of objects needed to understand and compare results between simulations.

An additional goal is to make it so that cohortIDs and cohortGroupIDs do not need to be consistent from year to year which will be useful for connecting other modules such as LandR. This can be accomplished either by including a full cohortID:cohortGroupID key for each year, or by excluding the cohortGroupID from the outputs completely.

NOTE: it is also desirable to save the disturbance matrix IDs used for each disturbance event.

Things to consider:
- What is useful to keep in the simList for easy access? How can this be kept as minimal as possible to prevent memory limits being reached?
- What file format is fastest to read & write to? How important is it for the data to be easily accessible?

cboisvenue:
Has this been dealt with in the recent CBMutils PR and the draft CBM_core PR#81?

PR #75, opened 2025-07-14
CBM outputs DB (DRAFT)

suz-estella:
Here's a brainstorm of how we could go about improving how we save output data from CBM_core (https://github.com/PredictiveEcology/CBM_core/issues/72). Marked as a draft as there's a lot that needs to be changed or discussed here before this is settled on but I think there's enough key pieces in place to review.

## Description

In this branch, the module no longer uses the `accumulateResults` event to create the `NPP` and `cbmPools` output tables in out simList. The module no longer requires that the use provide the "outputs" table as a project input. Instead, every year, the `cohortGroupKeep`, `cohortGroups`, and cbm_vars `state`, `flux`, and `pools` tables are dumped to parquet files that can later be queried as needed to get the result summaries required to create our output plots. The module has a new output - `cbmOutputsDB` - which is simply the path to the output directory containing the parquet files.

The queries to get the NPP and cbmPools tables are wrapped in functions (`simCohortGroupNPP` and `simCohortGroupPools`) that are cached to give easy and standardized access to the results. `CBMutils` could potentially contain these query functions to provide preestablished ways to read and summarize the output data for common purposes.

### Deciding on an output file format

This example reads & writes the outputs to parquet files. I chose this format because it is ideal for fast reading&writing to file while still having options to query the data without reading the entirety of each file as a single object. I think it would be wise to review and compare the pros & cons of a few different file formats (.qs, SQLite, e.g.) before settling on a choice. We can still settle on the _design_ of the module -- e.g. when are the outputs saved? Where to? -- without making a final choice on this.

If we find that there are certain results summaries that tend to be of particular interest we could consider having the module export these to CSV for easy access. I think this could be a good way to have both fast read&write of our data to a special format while still having some user friendly access to it.

### Implementing saving parameters

Saving all this data increases the workload of the module quite a bit since it's saving 5 potentially very large tables every year. This process definitely needs to be refined.

This brainstorm has 3 parameters set up that might be able to dictate how the data is saved:
- `.saveSpinup`: Save the tables after the spinup (default TRUE)
- `.saveAll`: Save all available data to file (TRUE) or only the data required for plotting & other key results tables (FALSE - the default). Currently, either way, this just saves all the available data in every table which is pretty inefficient. It would be a good to review what data needs to always be saved, and further, how to save it most efficiently (e.g. by rounding).
- `skipCohortGroupHandling`: I'm wondering if there's a way to have this parameter dictate whether all the data needs to be saved each year or just the updated data. For example, if CBM is handlling the cohort groups, then the cohort group key will only need to have new groups added to it each year; if it's not, then the entire key will need to be saved every year.

### Speeding up results queries and simpifying our plotting functions

the `sim$NPP` and `sim$cbmPools` tables are summaries of cohort group data for each year. The `simCohortGroupNPP` and `simCohortGroupPools` functions that replace these (in R/outputs-read.R) are useful for producing replicas of the tables our plotting functions expect, but I think we may benefit from replacing them with more direct queries. It would be simpler to write our plotting functions to work on results tables that summarize the data by pixel or year instead of by cohort group. Querying the results data directly this way may make data reading & manipulation for more efficient.

### DIFF
--- a/tests/testthat/test-3-integration_LandRCBM_RIA-small_2000-2002.R
+++ b/tests/testthat/test-3-integration_LandRCBM_RIA-small_2000-2002.R
@@ -21,8 +21,8 @@ test_that("Multi module: RIA-small with LandR 2000-2002", {
     SpaDES.project::setupProject(
 
       modules = c(
-        paste0("PredictiveEcology/Biomass_core@",    Sys.getenv("BRANCH_NAME")),
-        paste0("DominiqueCaron/LandRCBM_split3pools@run-with-CBM"),
+        paste0("PredictiveEcology/Biomass_core@", Sys.getenv("BRANCH_NAME")),
+        "PredictiveEcology/LandRCBM_split3pools@main",
         "CBM_core"
       ),
 


DominiqueCaron:
> We can still settle on the _design_ of the module -- e.g. when are the outputs saved? Where to? -- without making a final choice on this.

For when, `defineModule()` has the parameters `.saveInitialTime`, and `.saveInterval` which can be used here.

> * `.saveAll`: Save all available data to file (TRUE) or only the data required for plotting & other key results tables (FALSE - the default). Currently, either way, this just saves all the available data in every table which is pretty inefficient. It would be a good to review what data needs to always be saved, and further, how to save it most efficiently (e.g. by rounding).

I agree that the default should be only the data for plotting and essential for the query functions, and having an optional parameter to add more flexibilty if a user need additional objects (`.saveObjects`: a vector of the object to save?). I think the question "what are the queries we need/want?" would determine the outputs we want. I think that `cbm_var$pools` and `cbm_vars$flux` + a key table to link cohort groups to pixels (cohortGroupKeep, possibly only the column pixelIndex, cohortID, and cohortGroupID) + a key table to link cohort groups to species/age (cohortGroups) are definitely needed. Possibly, if this is not too "expensive", we could combine the two key tables to have: cohortID, pixelIndex, species, age, cohortGroup.

suz-estella:
All great feedback! I would like to address some general inefficiencies in the CBM_core code before returning to this as I think it would be a better standpoint from which to finish this work. Will circle back soon!

DominiqueCaron:
I looked at the sizes of the different objects and how they grow with simulation times, and it had brought some additional thoughts:

From the beginning, the objects that are the largest are `cbm_vars`, `spinupResult`, and `cbmPools`. If we write the pools to disk, we could remove `spinupResults` and `cbmPools` from memory.

The other object of interest is `cohortGroupKeep`, which keeps the history of cohort grouping across the simulation. This object grows in time, and if we think of doing long simulation (100 years or more), I suspect that it will become unmanageable. If we write to disk the key table discussed above, we could keep just the `cohortGroupID`, and `cohortGroupPrev` columns.

suz-estella:
> I looked at the sizes of the different objects and how they grow with simulation times, and it had brought some additional thoughts:
> 
> From the beginning, the largest objects are `cbm_vars`, `spinupResult`, and `cbmPools`. If we write the pools to disk, we could remove `spinupResults` and `cbmPools` from memory.
> 
> The other interesting object is `cohortGroupKeep`, which keeps the history of cohort grouping across the simulation. This object grows in time, and if we think of doing long simulations (100 years or more), I suspect that it will become unmanageable. If we write to disk the key tables discussed above, we could keep just the cohortID, pixelID, cohortGroupID, and cohortGroupPrev columns in memory.

I agree! My goal is to have is to that there's only data about "active" cohort groups and their current state at any given year. The rest can be dumped to file for review later.

PR #76, opened 2025-07-15
cohortDT 'ages' column renamed 'age'

suz-estella:
@cboisvenue as discussed: the cohortDT 'ages' column is made singular (one age per cohort) by dropping the 's'. @DominiqueCaron I imagine this requires an update on your end!


DominiqueCaron (approved):
I believe this should be fine on the `CBM_core` side, it can be merged!

I think that I set up `LandRCBM_split3pools` to work already, but I'll double-check.

Thanks

PR #80, opened 2025-07-28
LandR integration test branch update

suz-estella:
@DominiqueCaron I noticed that the `DominiqueCaron/LandRCBM_split3pools@run-with-CBM` branch is no more. I also see there's no `PredictiveEcology/LandRCBM_split3pools@development` branch. Would this be the correct fix to the integration test branch list?


DominiqueCaron:
Yes! That should be good. I forgot to update it.

PR #79, opened 2025-07-18
cbm_vars holds all interim data

suz-estella:
In this PR, the `cohortGroups` and `cohortGroupKeep` objects are no longer part of the module. Instead, the `cohortGroupKeep` has been moved to the new `cbm_vars` "key" table, and the spinup adds all cohort grouping variables (e.g. "spatial_unit_id", "gcids", etc) that were previously kept in `cohortGroups` to the `cbm_vars$state` table. The `cohortGroupID` column name is no longer used within the module: `row_idx` is the only column name that represents cohort group IDs for better consistency.

The `spinupResult` object has also been removed from the output objects. Later, I'd like to instead provide an option to save and review the output of the spinup via save parameters.

The next step will be looking to replace the use of tracking the full cohort group ID history in `cbm_vars$key` in favor of having this info saved during save events. For now, this info stays to keep the plot event working.

@DominiqueCaron: can `LandR_split3pools` be made work with this update? Let me know if you think we could get there. The key considerations I believe will be:
1. The `sim$spinupResult` is gone. Can `LandR_split3pools` edit `sim$cbm_vars` directly?
2. The `disturbance_type_id` column defining disturbance events for the year can now be found in `sim$cbm_vars$key` instead of `sim$standDT`.

----

`profvis` results running the SK managed forests 1985-2020:

| Branch | Event | RAM (GB) | time (s) |
| --------| ------- | ------| ----- |
| development | spinup | ~ +24, -9.7 | ~ 26 |
| development | annual_preprocessing | ~ +150, -176 | ~ 34.5 |
| development | annual_carbonDynamics | ~+185, -172 | ~ 51 |
| development | spades (total) | ~ +359, -349 | ~ 115 |

| Branch | Event | RAM (GB) | time (s) |
| --------| ------- | ------| ----- |
| suz-cbm_vars | spinup | ~ +11, -1.9 | ~ 17 |
| suz-cbm_vars | annual_preprocessing | ~ +36, -34 | ~ 6 |
| suz-cbm_vars | annual_carbonDynamics | ~ +64, -50 | ~ 19 |
| suz-cbm_vars | spades (total) | ~ +111, -86 | ~ 41 |

PR #81, opened 2025-08-15
spadesCBMdb replaces NPP and cbmPools

suz-estella:
An update to https://github.com/PredictiveEcology/CBM_core/pull/75 that improves how CBM_core's yearly data is written & read. This was made easier by https://github.com/PredictiveEcology/CBM_core/pull/79 which reduces the number of objects used by the module.

## Description

This update pairs with https://github.com/PredictiveEcology/CBMutils/issues/61 to introduce the SpaDES CBM outputs database to the **CBM_core** module. This PR has been marked as a draft until that PR is merged.

The `NPP` and `cbmPools` output objects are replaced by an output object called `spadesCBMdb` which is a path to a directory containing yearly data dumps of the `cbm_vars` tables to `.qs` files. The `accumulateResults` event no longer needs to occur. Instead, `CBMutils` contains functions that reads and summarizes the saved data, caches it, then passes the summaries to our plotting functions.

This solves the issue we have with our simulations requiring that the user provides the correct `outputs` table object in order to produce the listed module outputs that they require for plotting. This ensures that plotting will always be functional.

Having a special cache sub-directory in `spadesCBMdb` ensures that the main simulation cache doesn't return results from a previous run of the simulation sharing the same `cachePath`. The path to `spadesCBMdb` is always the same for simulations using a certain `outputPath`. To ensure the results are always current, each fresh run of the simulation will delete the directory to ensure the results (and the results cache) is fully overwritten.

### New parameters

- `.saveInitial`: Defaults to `start(sim)`.
- `.saveInterval`: Defaults to 1.
- `.saveAll`: default FALSE. Each year, the `cbm_vars$key`, `cbm_var$flux`, and `cbm_vars$pools` tables are saved to file. If `.saveAll = TRUE`, the `cbm_vars$parameters` and `cbm_vars$state` tables are also saved.
- `.saveSpinup`: default FALSE. Save the tables after the spinup. I have kept a test that sets this to TRUE and checks that the result is as we expect.

Plotting will only include years that had save events (as set with `.saveInitial` and `.saveInterval`).

`.saveAll` default data saving choices were guided by the plotting functions: `cbm_vars$pool` is required each year by `barPlot` to plot the change in pool proportions over time; `cbm_vars$flux` is required to plot NPP.

### Efficiency gains

Profile results running the SK test area, 30m res 28753x21700 masterRaster, 1985-1990:

With plotting:

| Branch | Format | RAM (MB) | time (s) |
| --- | --- | --- | --- |
| development | `NPP` & `cbmPools` | 383,524 | 705 |
| suz-CBMdb | .qs | 327,714 | 452 |

Without plotting:

| Branch | Format | RAM (MB) | time (s) |
| --- | --- | --- | --- |
| development | `NPP` & `cbmPools` | 367,189 | 574 |
| suz-CBMdb | .qs | 297,820 | 439 |

There are 2 reasons why our current `development` branch lags behind:
1. For runs without plotting: in development, the `NPP` and `cbmPools` objects are created regardless of whether they are being used anywhere. This update allows us to run the simulation, produce results, and leave the heavy lifting of summarizing the data for later when it's time to generate results.
2. Having the large tables `NPP` and `cbmPools` in our simList forces the module to cache them after every `annual_carbonDynamics` and `accumulateResults` event where they are modified (I believe): so instead of this data being saved once to file, they're actually being saved multiple times. In the `development` run, it was observed that ~2.6GB of data was being saved for every `annual_carbonDynamics` event; this was reduced to  ~1.6GB in the .qs run.

### What's next

Non-urgent: do we want to consider having the module save some useful data summaries to CSV for easy user access? This could happen either by default or we could add parameters that provide options to do this.

### DIFF
--- a/CBM_core.R
+++ b/CBM_core.R
@@ -14,8 +14,8 @@ defineModule(sim, list(
   citation = list("citation.bib"),
   documentation = list("README.txt", "CBM_core.Rmd"),
   reqdPkgs = list(
-    "data.table", "reticulate",
-    "PredictiveEcology/CBMutils@development (>=2.0.3.0013)",
+    "data.table", "reticulate", "qs",
+    "PredictiveEcology/CBMutils@development (>=2.1)",
     "PredictiveEcology/libcbmr"
   ),
   parameters = rbind(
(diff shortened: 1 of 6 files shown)


DominiqueCaron:
Wow this is really awesome @suz-estella ! 

My only concern is when we will want to do long simulation and/or experiments. Do you have an idea of how big the .qs files that are saved each year? I wonder if we want the option to only save at a certain interval or set the years we want to save. Let's say we simulate for >500 years, maybe saving every year a bit too much, especially at the time of summarizing over years, i.e., having to load >500 potentially large .qs files.

suz-estella:
@DominiqueCaron @camillegiuliano I've reintroduced the `.saveInitial` and `.saveInterval` parameters so that the user can decide which years are saved. The `plot` event will only plot years that had a save event.

PR #95, opened 2026-01-22
spinupSQL input object replaced by cbm_defaults_db; cbm_exn_dir input added

suz-estella:
Follows: https://github.com/PredictiveEcology/CBM_defaults/pull/26

Input object changes:
1. `cbm_defaults_db`: Optional. Path to CBM defaults SQLite database. Replaces the `spinupSQL` input which was a query of the "spatial_unit" and "spinup_parameters" tables of the CBM defaults DB.
2. `cbm_exn_dir`: Optional. Path to CBM-EXN parameters directory.

A great benefit to this is that the module does not require the user to provide either of these inputs. `libcbmr` pulls the default parameters from the Python install of libcbm (`~/.virtualenvs/r-spadesCBM/Lib/site-packages/libcbm/resources`) and will use these unless the user decides to provide custom paths to these resources. In our case, when running the module with CBM_defaults, we will always be passing these paths from the resources created in our project directory, even if CBM_defaults hasn't changed any of the parameters from the defaults.

Adding the `cbm_exn_dir` input introduces the ability to customize any of the CBM-EXN parameters, which we didn't have a way to do before.

While using `cbm_defaults_db` instead of `spinupSQL` adds some extra work to query the DB, it provides clarity to what the source of this data is, and gives space for a straightforward way to customize these parameters outside of the module as needed: the user can edit parameter files directly in any way that is suitable to their purpose. This makes it easy to our feed custom parameters to functions that use any of these resources to create summary tables or plots, because they can be written to work with files in a predictable format.

This sets us up well to move towards using CBM4 which will also make use of these paths to resources. It may use the parameters in different ways but the inputs to the module will not need to be changed.


cboisvenue (approved):

cboisvenue:
@camillegiuliano @DominiqueCaron can you look at this PR to see if it works for you?

camillegiuliano (approved):

DominiqueCaron (approved):

DominiqueCaron:
Ok, looks good. I will need to double-check if I need to make some changes to LandRCBM.

PR #96, opened 2026-02-04
poolsToPlot parameter removed

suz-estella:
I noticed this parameter isn't being used anywhere. Perhaps it got lost somewhere when we updated our plotting. Is this something we want to link back in? Or do we just need to remove it as a module parameter? If the latter, this PR accomplishes that.

### DIFF
--- a/CBM_core.R
+++ b/CBM_core.R
@@ -43,9 +43,6 @@ defineModule(sim, list(
       "emissionsProductsCols", "character", c("CO2", "CH4", "CO", "Emissions"), NA_character_, NA_character_,
       desc = "A vector of columns to return for emissions and products"),
     defineParameter(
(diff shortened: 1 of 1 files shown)


camillegiuliano:
My memory is foggy here, but I think the general idea was to incorporate this parameter so that what is now the total carbon map could be changed to plot different pools combinations rather than only total carbon. But it was never implemented. I think it's safe to remove, I don't really see a point in having it, a user could just run the plot function with whatever pools they'd want to visualize with the resulting `sim` anyway.

camillegiuliano (approved):

cboisvenue (approved):

cboisvenue:
I did use the different combinations of pools before any of you got here. I don't think we need it anymore. Nice catch susan.

PR #100, opened 2026-02-17
`standDT` has the 'admin_name' and 'eco_id' columns instead of 'spatial_unit_id'

suz-estella:
This is a part of a few module updates to make things more compatible with CBM4. One thing that is notable in CBM4 is that it's going to require the administrative boundary full name (e.g. "Saskatchewan") and the ecozone ID to run instead of just the CBM spatial unit ID. 

### Description

**Updates:**
1. `cbmEXN_spinup` can accept a table with 'admin_name' and 'eco_id` or the 'spatial_unit_id' already preset
2. `standDT` documentation asks the user to provide the 'admin_name' and 'eco_id' columns instead of 'spatial_unit_id'

Since `cbmEXN_spinup` can still accept the 'spatial_unit_id' that we are currently passing from **CBM_dataPrep** this PR will not affect our current simulations.

This is a move towards having the responsibility of getting the spatial unit ID from **CBM_dataPrep** to **CBM_core**. This occurs during a step where the CBM defaults db is already being read to retrieve the spinup parameters, so I think this makes good organizational sense. I think it also logically makes sense to start thinking of the spatial unit ID as an "internal ID" that the user doesn't need to necessarily be aware of or set correctly during data prep.

Following this I will PR:
1. An update to **CBM_vol2biomass** that uses `userGcLocations` instead of `userGcSPU` (https://github.com/PredictiveEcology/CBM_vol2biomass/issues/37): a table with `admin_abbrev` and `eco_id` instead of `spatial_unit_id`. This will allow us to no longer need to set the `spatial_unit_id` in **CBM_dataPrep**.
2. An update to **CBM_dataPrep** that adds the `admin_name` and removes the `spatial_unit_id` from the standDT table.

### Naming conventions

Interested in any feedback here:

I've opted to rename the `ecozone` field to `eco_id`. This is because in the CBM4 world, there is the `eco_boundary_id` (the ecozone ID) and the `eco_boundary` which is the ecozone name. My thinking here is that 'ecozone' is too generic to know which one you are working with.

If we wanted to standardize how we provide the admin & ecozone columns across our modules, I might suggest:

Ecozone columns:
- `eco_id`: Ecozone ID; same as the CBM `eco_boundary_id`
- `eco_name`: Ecozone name; in CBM4 this is the `eco_boundary`

Admin columns:
- `admin_abbrev` **: Admin abbreviation ("SK"); not used by CBM. Used by **CBM_vol2biomass** to retrieve Boudewyn parameters where this is called the `juris_id`.
- `admin_name`: Admin name ("Saskatchewan"); in CBM4 this is the `admin_boundary`

** _Should we rename `admin_abbrev` `admin_id`_? I note that in the CBM defaults DB, there is a numeric `admin_boundary_id` that may be used internally by **CBM_core** but not elsewhere. If we're not worried about potentially getting those confused I would maybe opt for that.

### DIFF
--- a/CBM_core.R
+++ b/CBM_core.R
@@ -54,12 +54,13 @@ defineModule(sim, list(
   ),
   inputObjects = bindrows(
     expectsInput(
-      objectName = "standDT", objectClass = "data.table", sourceURL = NA,
+      objectName = "standDT", objectClass = "data.table",
       desc = "Table of stand attributes. Stands can have 1 or more cohorts.",
       columns = c(
-        pixelIndex      = "Stand ID",
-        area            = "Stand area in meters",
-        spatial_unit_id = "CBM-CFS3 spatial unit ID",
+        pixelIndex = "Stand ID",
+        area       = "Stand area in meters",
+        admin_name = "Canada province or territory name",
+        eco_id     = "Canada ecozone ID",
         historical_disturbance_type = "Historic CBM-CFS3 disturbance type ID. Defaults to the 'historical_disturbance_type' parameter",
         last_pass_disturbance_type  = "Last pass CBM-CFS3 disturbance type ID. Defaults to the 'last_pass_disturbance_type' parameter"
       )),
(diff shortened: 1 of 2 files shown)


cboisvenue:
Nice work Susan.
Naming convention: I have been trying to match the CBM-CAT naming convention since the beginning. They keep changing it, but it seems to be stabilizing (?). I suggest the name convention that had the least chance of being confusing.

cboisvenue (approved):

camillegiuliano (approved):

PR #102, opened 2026-02-23
integration test for LandR-CBM simplified

### LINKED ISSUE #101: LandR-CBM integration test failure checking species IDs
This line in the integration test presents an error:

https://github.com/PredictiveEcology/CBM_core/blob/8d17c56ba40d50a6a065c376f38b9c8b28bc9719/tests/testthat/test-3-integration_2-LandRCBM_RIA-small.R#L90

It seems as though the CBM species IDs that are expected to be 16 are 15 instead. For reference, species ID 15 = "Lodgepole pine", 16 = "Shore pine"

It's maybe worth mentioning that in CBM4 we won't need to provide the CBM species ID with out cohorts/curves and this ID _may_ not be readily available any output tables at all. Currently it's just being used by CBM EXN to determine if the species is HW or SW but CBM4 doesn't require this (redundant) bit of info when we're passing it our carbon increments.

suz-estella:
Closes #101 

I think it's good that we have reviewed this changed result and have decided we now have the correct species IDs making their way to the module.

For the update to the integration test: I think it might be best to take this moment to simplify it instead of setting ourselves to have to keep maintaining it over time. This PR removes specific validations to instead just check that the simulation runs without error. It also shortens the simulation time span from 22 years to 3.

The other integration test is already similar to this: https://github.com/PredictiveEcology/CBM_core/blob/development/tests/testthat/test-3-integration_1-SK-small.R

_Why I recommend this_: I'm starting to think that with **CBM_core**, we _mostly_ want integration tests to catch where we may have introduced changes that make the modules incompatible since they were developed to work with it. The other individual modules we are including in the test simulations (e.g. **CBM_dataPrep**, **CBM_vol2biomass**, **LandRCBM_split3pools**) would be responsible for containing more detailed tests checking for specific things such as looking for expected values in the results. I think these checks are more useful there because they exist where the modules are being developed. Aside from major updates, **CBM_core** should really be the module that is being developed the least frequently.

@DominiqueCaron maybe this integration test with the lengthier validation checks could be a part of **LandRCBM_split3pools** instead? That way, when it's time to update it to work with CBM4, the tests checks can be updated to fit that new structure as well. For example, the `cbm_vars` object isn't a part of the CBM4 version of the module.

### DIFF
--- a/tests/testthat/test-3-integration_2-LandRCBM_RIA-small.R
+++ b/tests/testthat/test-3-integration_2-LandRCBM_RIA-small.R
@@ -9,7 +9,7 @@ test_that("Multi module: RIA-small with LandR 2000-2002", {
 
   # Set up project
   projectName <- "integration_LandRCBM_RIA-small_2000-2002"
-  times       <- list(start = 2000, end = 2021)
+  times       <- list(start = 2000, end = 2002)
 
   simInitInput <- SpaDEStestMuffleOutput(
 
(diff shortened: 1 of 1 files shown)


cboisvenue (approved):

DominiqueCaron (approved):
Agreed! 

I believe that the longer simulation time was to test the correct association of cohorts between timesteps after a dispersal event (every 10 years by default in LandR). This test certainly belong more in `LandRCBM_split3pools`.

PR #105, opened 2026-04-20
gcMeta drops requirement for CBM species_id column

suz-estella:
I've got confirmation that we do not need our cohorts to be linked to a specific CBM species ID for CBM-EXN to work.

In CBM-EXN's [species.csv](https://github.com/cat-cfs/libcbm_py/blob/main/libcbm/resources/cbm_exn/species.csv), species are linked to a `forest_type_id` (SW or HW) and a `genus_id` that _would_ only ever be used in vol to biomass conversions in CBM. Since we provide our own carbon increments, using any CBM species ID that matches the HW/SW of our cohorts will produce the same results.

This update sets the `species` column for our cohort groups on-the-fly based on the value in the `sw-hw` column. Running our module SK-small test confirms that this doesn't have any affect on the output numbers.

Other species metadata (such as the species name, genus, etc) can still be reviewed in the `cbm_vars$state` and `gcMeta` tables.

This simplification will be useful to us because we will no longer need to maintain the `CBM_speciesID` column of [LandR::sppEquivalencies_CA](https://github.com/PredictiveEcology/LandR/blob/development/data-raw/sppEquivalencies_CA.csv) going forward. CBM4 does not require us to provide CBM species IDs as an input, but if it did, it would require the IDs in the CBM defaults database instead of the EXN species.csv table, meaning that the `LandR::sppEquivalencies_CA$CBM_speciesID` column would not be usable. Dropping this step makes the transition to CBM4 easier.

I will follow this by PRs for updates to **CBM_dataPrep** and **LandRCBM_split3pools** to no longer match for this column. However, I think we should wait until our next merge of our modules to the main branches before removing this column in LandR (I've kept note of this).


suz-estella:
re GHA fail: there's some sort of R package dependency resolution error happening only on Linux. Not sure what this is yet but I'll keep an eye on it. I believe the windows test pass is sufficient to merge.

cboisvenue (approved):

camillegiuliano (approved):

PR #107, opened 2026-05-01
Documentation updates

suz-estella:
A few module documentation updates, the main thing being that I updated the overview to mention and link to the the `libcbmr` R package and removed broken links to the CBM-EXN documentation.

Considering if it might make sense to keep a copy of the rendered (or un-rendered - a copy of the raw files) CBM-EXN documentation in the `libcbmr` repo in which case the documentation will be linked to here. Will think on this more.


cboisvenue (approved):

camillegiuliano (approved):

camillegiuliano:
I do think we should keep a copy of the cbm-exn documentation somewhere, at least while we are still using it. Even if it's just the PDF version you got from Scott, I'd say that's good enough.

